PR #230, opened 2026-07-16

phelma:

- main + pr workflows: loop-guard, check/test, prerelease, `release` environment gate
- Docker Hub credentials load made lazy (`dynamic` block, nil when the tree
  is locked) so PR checks run without unlocking git-crypt — the one
  deliberate change inside the `image` namespace

  `lib/version.rb` logic.
  carried `[skip ci]`, so this matches. Updates ship with the next
  human-triggered release.
  improvement.
- The `release` job pulls `main` at approval time, so a delayed approval
  releases main as it stands then, not the SHA this run tested — parity with
  the old `release.sh` (which also pulled; `prerelease.sh` did not, so the
  prerelease job has no pull).
  push-back commit and nothing for the loop-guard to catch; `skip-ci-check`
  stays as family-uniform belt-and-braces.
  supplies secrets/context.

Disabling the CircleCI project and deleting the `CircleCI` deploy key are
deferred to the end-of-migration sweep.


Gemfile.lock
```diff
@@ -206,7 +209,7 @@ GEM
     unicode-display_width (3.1.4)
       unicode-emoji (~> 4.0, >= 4.0.4)
     unicode-emoji (4.0.4)
-    uri (0.13.2)
+    uri (1.1.1)
 
 PLATFORMS
```
phelma, line 216: 🟡 **Standards — plan conformance gap (in-scope)**

§4.5 requires that *after any lockfile update* the `PLATFORMS` section "retains the generic `ruby` and `x86_64-linux` platforms and contains no operator-local platforms (e.g. `arm64-darwin-*`)", and prescribes the fix explicitly: `bundle lock --add-platform ruby x86_64-linux` and `bundle lock --remove-platform <darwin platform>`.

This PR updates the lockfile, but `PLATFORMS` still reads:

```
arm64-darwin-21
arm64-darwin-22
arm64-darwin-23
arm64-darwin-25
x86_64-linux
```

— four operator-local darwin platforms and **no generic `ruby` entry**. `arm64-darwin-25` matches this migration host (Darwin 25). The section is unchanged from `main`, so the drift pre-dates the PR; the plan nonetheless frames the check as a post-update obligation, not a no-change-required condition. This PR's own resolution also added a platform-specific `ffi (1.17.4-arm64-darwin)` alongside `ffi (1.17.4-x86_64-linux-gnu)`, so darwin-specific gem variants are now locked in by this change.

**Impact**: Working but wrong-shaped — `x86_64-linux` is present so the runner resolves correctly (the `test` job is green), and the missing generic `ruby` platform has no effect on CI today. The cost is fleet-shape divergence: this is the exact condition §4.5 tells the next ~100 repos to normalise, so leaving it silently sets the family precedent the plan was written to prevent.

**Suggestion**: `bundle lock --add-platform ruby` and `bundle lock --remove-platform arm64-darwin-21 arm64-darwin-22 arm64-darwin-23 arm64-darwin-25`, then re-verify with the §5 step-4 clean-env install. If keeping the darwin platforms is a conscious call, record it in the PR's deliberate-decisions list so the divergence is explicit rather than silent.

Rakefile
```diff
@@ -203,9 +214,11 @@ namespace :image do
...
     t.repository_url = 'infrablocks/alpine-aws'
 
-    t.credentials = YAML.load_file(
-      'config/secrets/dockerhub/credentials.yaml'
-    )
+    t.credentials = dynamic do
+      YAML.load_file('config/secrets/dockerhub/credentials.yaml')
+    rescue Psych::SyntaxError
+      nil # tree locked: skip Docker Hub auth (base image is public)
+    end
```
phelma, line 221: 🔵 **Correctness / Safety / Code Quality — PLAN CONCERN** (§4.4 step 2a specifies this snippet verbatim; flagged independently by three lenses)

Two separate points, one of which I verified empirically.

**1. The plan's stated rationale for `dynamic` is wrong — but the change is still needed.** §4.1 justifies this by claiming the eager `YAML.load_file` "evaluates at Rakefile *load* time, so any rake invocation in a locked tree dies with a `Psych::SyntaxError`". It does not. `RakeFactory::TaskSet` invokes the configuration block from inside the *task body* — `TaskDefinition#resolve_block` builds a lambda that `Task#define_task` calls within the rake task's action. I confirmed this against rake_factory 0.34: with an eager, raising `t.credentials`, `rake -T` lists all six `image:*` tasks without error. What actually fails is `image:prepare`/`image:build` at *task-run* time, which still breaks `test:integration` on every PR — so the rescue is genuinely load-bearing and the diff is right to include it. Only the reasoning needs correcting.

That said, it means the `dynamic` wrapper (and the `require 'rake_factory/kernel_extensions'` it needs) is doing less than the plan thinks: the `rescue` alone is what makes the locked tree survive. A plain `t.credentials = begin ... rescue Psych::SyntaxError; nil; end` would behave identically. Worth settling before it is replicated across ~100 repos.

**2. The rescue uses an exception type as a proxy for "the tree is locked".** To be fair to the current code, I checked the obvious worry and it does not hold: a file with the real git-crypt magic header (`\x00GITCRYPT\x00` + binary) raises exactly `Psych::SyntaxError` ("control characters are not allowed at line 1 column 1"), so the rescue is correctly targeted for the locked case. The real cost is at the other end: in the *unlocked* `prerelease`/`release` jobs, a genuinely malformed `credentials.yaml` is indistinguishable and also yields `nil`. `rake_docker` does `Docker.authenticate!(t.credentials) if t.credentials` (`build.rb:37`, `push.rb:21`), so it silently skips auth and attempts an unauthenticated push. This is bounded — `rake_docker`'s `Output.print` raises on `errorDetail`, so the job fails loudly rather than going green with nothing published — but it surfaces as an opaque registry denial rather than "your credentials file is broken", and only *after* the version tag has been pushed.

**Suggestion**: No change in this PR — the plan is the spec. For the family template, test the condition directly rather than inferring it from an exception class, e.g.

```ruby
t.credentials = dynamic do
  path = 'config/secrets/dockerhub/credentials.yaml'
  # tree locked: skip Docker Hub auth (base image is public)
  next nil if File.binread(path, 10) == "\x00GITCRYPT\x00"

  YAML.load_file(path)
end
```

At minimum, `warn` in the rescue so the log records that Docker Hub auth was skipped and why.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+# Every run publishes an rc image and tags the repo; serialise so
+# close-together merges cannot race on version tags. queue: max keeps every
+# queued run (the default keeps only the newest pending), so a slow release
+# approval delays later prereleases instead of cancelling them.
+# 'queue: max' — GA 2026-05-07: https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/
+concurrency:
+  group: main
+  cancel-in-progress: false
+  queue: max
```
phelma, line 18: 🔵 **Correctness / Safety — PLAN CONCERN** (§4.2 mandates this concurrency block, `queue: max` included)

Not asking for a change — asking for one verification step before merge.

This PR installs `main.yaml` and deletes `.circleci/config.yml` in the same commit, so at merge the repo has exactly one pipeline and no fallback. Nothing has evaluated `main.yaml` against the Actions workflow schema: `pr.yaml`'s green checks say nothing about it, and §5's verification is `YAML.load_file`, which proves YAML syntax only. I confirmed `gh api repos/infrablocks/docker-base-alpine-aws/actions/workflows` currently lists only `Pull Request` — expected, since registration reads the default branch, so that is *not* evidence of a fault, but it does mean `queue: max` stays unvalidated until after merge.

`queue: max` is a very recent concurrency key (the comment dates it 2026-05-07). Two failure modes: if it is rejected, an unrecognised key in `concurrency` is a **whole-workflow parse failure** — no `skip-ci-check`, no `check`, no rc, no release, presenting as a workflow annotation that is easy to miss rather than a red check. If it is silently ignored, the default queue behaviour cancels pending runs, the exact opposite of the comment's intent.

**Impact**: The first push to `main` after merge is the first time this file is ever evaluated, at which point CircleCI is already gone. Recovery is a revert, so this is a detection problem rather than an unrecoverable one.

**Suggestion**: Before merging, confirm the file is accepted — `actionlint` over `.github/workflows/` (worth adding to `test:code:check` family-wide, since it would also cover the other ~100 repos), or a scratch-repo push confirming a `concurrency` block with `queue: max` is accepted. If `queue` cannot be confirmed for this org's plan tier, drop the key: `group: main` + `cancel-in-progress: false` still serialises runs, and the queue-depth nuance is not worth risking the whole pipeline on.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+        continue-on-error: true
+        run: ./go "slack:notify[${{ job.status }}]"
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+  release:
+    needs: [prerelease]
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    environment: release
```
phelma, line 120: 🔵 **Safety — PLAN CONCERN** (§4.2's concurrency block is plan-mandated and its comment documents the queueing intent)

The plan's comment says "a slow release approval delays later prereleases instead of cancelling them". The effect is broader than that, and the comment is worth correcting even if the config stays.

`concurrency` is declared at **workflow** level, so a run occupies group `main` for its entire lifetime — including while `release` sits waiting on the `release` environment's required reviewers. Environment approvals wait up to 30 days. While a release is pending, subsequent merges to `main` do not run `check` or `test` either; they queue. Under CircleCI the `hold` job blocked only its own workflow — other pushes built independently — so this is a genuine behavioural change, and it is not on §1's documented-hazards list.

**Impact**: The realistic scenario is a release nobody got round to approving on a Friday: `main` accumulates merged commits with zero CI feedback for days, and queued runs emit no signal, so nothing alerts anyone that the pipeline has stalled. Recovery is cheap once noticed (reject the pending deployment and the group frees) — the problem is detection, not repair.

**Suggestion**: No change required at cutover. For the family: either scope concurrency to the jobs that actually race on version tags rather than the whole workflow, or set a wait-timer on the `release` environment. At minimum, reword the comment — it currently understates what the queue blocks, and that comment will be copied into ~100 repos.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Install tools
+        uses: infrablocks/github-actions/asdf_install@v1
+      - name: Ensure docker-compose is available
+        run: |
+          if ! command -v docker-compose >/dev/null 2>&1; then
+            sudo curl -fsSL \
+              "https://github.com/docker/compose/releases/latest/download/docker-compose-linux-x86_64" \
+              -o /usr/local/bin/docker-compose
+            sudo chmod +x /usr/local/bin/docker-compose
+          fi
```
phelma, line 39: 🔵 **Security / Safety — PLAN CONCERN** (§4.1/§4.2 specify this step verbatim; duplicated at `main.yaml:65-71`)

The step downloads `releases/latest` — a mutable tag — over the network, writes it to `/usr/local/bin` as root, and marks it executable, with **no version pin and no checksum or signature verification**. Whatever Docker publishes as "latest" at run time is what executes.

Worth sharpening for this repo specifically: there is no `spec/dependencies.yml` and no `dependencies:test:*` tasks in the Rakefile — `test:integration` depends only on `image:build`, and `spec/image_spec.rb` drives serverspec's docker backend directly. **Nothing here will ever invoke `docker-compose`.** The plan calls the guard "a no-op" for such repos, and I want to be precise about that: it is a no-op in *usage*, not in *effect*. The step still runs (it completed successfully on this PR's `test` job), and on a runner lacking the standalone v1 binary it installs an unpinned, unverified root-owned binary for no benefit — pure added supply-chain surface, plus a network dependency in the path of a job that gates `prerelease` and `release` via `needs`.

**Impact**: An unverified third-party binary is trusted on every run of a repo that never uses it; a compromised or regressed upstream release lands automatically with no pin to roll back to. Latent non-determinism family-wide once a repo with real compose dependencies adopts the same shape — a future compose release could change test behaviour with no commit.

**Suggestion**: Keep the one-family-shape decision, but pin the version and verify a checksum rather than tracking `latest` — that removes the non-determinism without reintroducing a variant. Note the hard-coded `x86_64` is correct for `ubuntu-latest` today but would silently install the wrong binary on an arm64 runner.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,61 @@
...
+        if: ${{ !cancelled() }}
+        continue-on-error: true
+        run: ./go "slack:notify[${{ job.status }}]"
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+  merge-pull-request:
+    needs: [check, test]
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
```
phelma, line 52: 🔵 **Security — PLAN CONCERN** (D3: "No update-type filter, also parity")

Dependabot PRs auto-merge on green checks with no update-type or ecosystem filter, and `.github/dependabot.yml` covers both `bundler` and `docker` (daily) — so a compromised gem release or a poisoned `alpine` base-image tag merges to `main` unreviewed.

Being precise about blast radius, since it is easy to overstate: the PR run itself holds no secrets (read-only token, no git-crypt unlock), and the `GITHUB_TOKEN` merge does not trigger `main.yaml` — so nothing publishes at merge time. The exposure is **deferred**: the compromised dependency sits on `main` until the next human-triggered release, where `prerelease`/`release` run it in a tree with `ENCRYPTION_PASSPHRASE` unlocked, Docker Hub credentials on disk, and `contents: write`.

**Impact**: A malicious dependency update reaches `main` without human review and later executes in the one job that holds every secret this repo has.

**Suggestion**: No change at cutover — documented parity, and the deferral genuinely reduces the immediate risk. For the hardening pass: gate auto-merge on `steps.metadata.outputs.update-type` from `dependabot/fetch-metadata`, allowing `semver-patch`/`semver-minor` and leaving major bumps to a human.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+          fi
+
+  check:
+    needs: [skip-ci-check]
+    if: needs.skip-ci-check.outputs.should_skip_ci != 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install tools
+        uses: infrablocks/github-actions/asdf_install@v1
```
phelma, line 45: 🔵 **Security — PLAN CONCERN** (D6; §1 lists this explicitly as "not drift to flag")

`infrablocks/github-actions/asdf_install@v1` is a moving major-version tag, not a commit SHA, and it runs in every job in both workflows — including `prerelease`/`release`, where git-crypt is unlocked and `contents: write` is held.

Recording the tradeoff rather than disputing it: this is a self-owned action in the same org, so no trust boundary is actually being crossed, and SHA-pinning 103 repos would create a dependency-bump treadmill worse than the risk it removes. The residual risk is narrow but real — a compromise of `infrablocks/github-actions`, or a maintainer account silently repointing `v1`, propagates into every repo in the fleet at once with no per-repo change to review.

**Impact**: Single upstream compromise reaches the whole fleet with no pinning to slow it down.

**Suggestion**: No change here. The proportionate fleet-wide mitigations are tag protection on `infrablocks/github-actions` and requiring review on `v1` retags — cheaper and more effective than pinning.

Rakefile
```diff
@@ -142,51 +134,70 @@ namespace :library do
...
-  desc 'Prepare CircleCI Pipeline'
+  desc 'Prepare GitHub Actions pipeline'
   task prepare: %i[
-    circle_ci:env_vars:ensure
-    circle_ci:checkout_keys:ensure
-    circle_ci:ssh_keys:ensure
-    github:deploy_keys:ensure
+    github:secrets:ensure
+    github:environments:ensure
   ]
```
phelma, line 201: 🔵 **Correctness — PLAN CONCERN** (§4.4 step 6 specifies the `ensure` tasks)

`pipeline:prepare` depends on `github:secrets:ensure` and `github:environments:ensure`. In rake_github 0.17.0 the `ensure` tasks are **not** create-or-update — they invoke `destroy` and *then* `provision` (`tasks/secrets/ensure.rb:21-22`, `tasks/environments/ensure.rb:21-22`).

So re-running provisioning opens a window in which `ENCRYPTION_PASSPHRASE` does not exist, and deletes and recreates the `release` environment. A `prerelease`/`release` job whose `Unlock git-crypt` step lands in that window fails; a `release` job sitting in the approval queue has its gate destroyed underneath it.

**Impact**: The PR description and §7 both call the task idempotent and say "re-runs are fine". That is true of the end state but not of the transition — a re-run concurrent with an in-flight main run can fail that run or disturb a pending release approval. Small window, real TOCTOU, and operators are being told it is safe.

**Suggestion**: rake_github also defines `provision` tasks which *are* create-or-update (`create_or_update_environment`, `environments/provision.rb:27`). Either point `pipeline:prepare` at those, or note in the runbook that provisioning must not run while a main workflow is in flight.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Install secrets tools
+        run: sudo apt-get update && sudo apt-get install -y git-crypt gnupg
+      - name: Unlock git-crypt
+        run: ./go git_crypt:unlock_with_encrypted_gpg_key
+        env:
+          ENCRYPTION_PASSPHRASE: ${{ secrets.ENCRYPTION_PASSPHRASE }}
+      - name: Set CI git author
+        run: ./go repository:set_ci_author
+      - name: Prerelease
+        run: ./go "version:bump[rc]" && ./go image:publish
```
phelma, line 103: 🔵 **Safety — PLAN CONCERN** (D5: `version:*` logic is deliberately untouched; related to §1's documented tag-then-publish hazard)

Adjacent to the documented hazard, via a different route worth recording. `version:bump[rc]` calls `repo.push('origin', 'main', tags: true)` — it pushes the **branch** as well as the tag. The `prerelease` job deliberately has no `git pull` (correct, mirroring `prerelease.sh`), so it pushes `main` at the SHA it checked out. If a `release` job — which *does* pull — has advanced `origin/main` while a later run sat queued, that queued prerelease pushes a stale `main` and git rejects it as non-fast-forward. Since `git push` is not atomic across refs by default, the tag can still land while the push exits non-zero, orphaning a version tag with no image behind it.

**Impact**: Low probability, low damage — an orphan tag plus a confusing red prerelease. Noting it because the `queue: max` design deliberately lengthens the staleness window that makes it possible, so the two decisions interact.

**Suggestion**: No change at cutover — D5 territory, and the same shape existed on CircleCI. For the post-migration list alongside tag-then-publish ordering: the docker family versions by tag and has no reason to push the branch at all, so `version:bump` could push tags only (`repo.push('origin', tags: true)`).

phelma (commented):
## Code Review: #230 - Migrate pipeline from CircleCI to GitHub Actions

**Verdict:** COMMENT

This is an unusually faithful implementation of the Variant B family plan: both workflow files match §4.1/§4.2 essentially verbatim, the Rakefile edits match §4.4 step-for-step, Slack channel IDs match §3's table exactly, the GPG key move is a pure rename, all §4.7 deletions are present, `.gitattributes`/`src/`/`spec/`/`lib/version.rb` are untouched per §4.8, and the commit message follows §6. The `pr.yaml` checks are green on the real runner (`check` 4m57s, `test` 5m39s, `merge-pull-request` correctly skipped), which validates the D8 headline risk — the native docker daemon — for this repo.

One in-scope conformance gap was found (the `Gemfile.lock` `PLATFORMS` section, which §4.5 explicitly requires normalising after a lockfile update) and one gap the plan does not cover at all (the deleted CircleCI/GitHub credentials remain valid — `git rm` is not revocation). Everything else the lenses raised challenges a decision the plan or PR description documents as deliberate; those are recorded below as **plan concerns** so a human can revisit the *plan*, and they do not block this diff.

### Verified During Review

Two lens claims were checked empirically against the `rake_factory` 0.34 source and a synthetic git-crypt-locked fixture, rather than taken on inspection:

- ✅ **The `dynamic` rescue catches the right exception.** A file with the real git-crypt magic header (`\x00GITCRYPT\x00` + binary) raises exactly `Psych::SyntaxError` (*"control characters are not allowed at line 1 column 1"*). Speculation that a locked file might raise `ArgumentError` or a bare `Psych::Exception` and take down every PR check is **disproven** — the rescue is correctly targeted.
- ⚠️ **The plan's stated rationale for `dynamic` is factually wrong, but the change is still load-bearing.** §4.1 claims the eager `YAML.load_file` means "any rake invocation in a locked tree dies at Rakefile *load* time". It does not: `RakeFactory::TaskSet` invokes the configuration block from inside the *task body* (`TaskDefinition#resolve_block` → `Task#define_task`), so `rake -T` succeeds against an eager, raising `t.credentials`. What actually fails is `image:prepare`/`image:build` at *task-run* time — which still breaks `test:integration` on every PR, so the rescue is genuinely required. Only the reasoning needs correcting, not the code. Detail inline.

### Cross-Cutting Themes

- **The `dynamic`/`rescue Psych::SyntaxError` credentials block** (flagged by: correctness, safety, code-quality) — all three independently landed on the same property: the rescue uses an exception type as a proxy for "the tree is locked", so in the *unlocked* `prerelease`/`release` jobs a genuinely malformed credentials file also yields `nil`, and `rake_docker`'s `Docker.authenticate!(t.credentials) if t.credentials` then attempts an unauthenticated push. The plan specifies this snippet verbatim, so it is a plan concern — but three lenses converging makes it the strongest candidate for a family-wide revisit.
- **`main.yaml` is never exercised before it becomes the only pipeline** (flagged by: correctness, safety) — this PR installs `main.yaml` and deletes `.circleci/config.yml` in one commit. `pr.yaml`'s green checks prove nothing about `main.yaml`, and §5's verification is `YAML.load_file`, which checks YAML syntax, not the Actions workflow schema. I confirmed `main.yaml` is not yet registered (`gh api .../actions/workflows` lists only `Pull Request`) — expected, since workflow registration reads the default branch, so this is not evidence of a fault, but it does mean `queue: max` is genuinely unvalidated until after merge.

### Tradeoff Analysis

- **Family uniformity vs surface reduction (the docker-compose guard)**: this repo has no `spec/dependencies.yml` and no `dependencies:test:*` tasks — `test:integration` depends only on `image:build`, so nothing here will ever invoke `docker-compose`. The plan knowingly accepts a dead step to keep one family shape ("fewer variants"), which is a defensible call. Security's counterpoint stands on the record: the step is a no-op in *usage*, not in *effect* — it still runs, and on a runner without the standalone binary it installs an unpinned, unverified root-owned binary. The guard step did complete successfully on this PR's run. Recommendation: keep the shape, pin the version family-wide.
- **Parity vs hardening (dependabot auto-merge, `asdf_install@v1`)**: D3 and D6 deliberately preserve CircleCI-era behaviour. Worth noting the blast radius is narrower than it first looks — the PR run holds no secrets and the `GITHUB_TOKEN` merge triggers no release — so the exposure is deferred to the next human-triggered release rather than immediate. Correct call for a cutover; both belong on the hardening list.

### Strengths

- ✅ Near-verbatim conformance to §4 in both directions — no required change missing (bar `PLATFORMS`), and nothing changed beyond the plan. No rubocop autocorrect noise or drive-by refactoring crept in.
- ✅ `pr.yaml` uses `on: pull_request`, not `pull_request_target` — fork PRs that build PR-authored Dockerfiles get a read-only token and no repository secrets. The correct trigger choice.
- ✅ Untrusted context is handled correctly throughout: `github.event.head_commit.message` and `github.event.pull_request.html_url` are bound via `env:` and referenced as `"$VAR"` inside quoted shell, closing the classic GHA script-injection vector. Only `job.status` — a trusted enum — is interpolated directly.
- ✅ Least-privilege token scoping: workflow-level `contents: read` default-deny, with `contents: write` only on `prerelease`/`release` and `pull-requests: write` only on `merge-pull-request`.
- ✅ `ENCRYPTION_PASSPHRASE` is provisioned to the Actions store only (no `dependabot: true` mirror) and referenced only in `main.yaml` — dependabot runs cannot reach it. The lazy credentials block is what makes this hold, and the reasoning is sound.
- ✅ `fetch-depth: 0` is on exactly the jobs that resolve `latest_tag` (`test`, `prerelease`, `release`) and correctly omitted from `check`.
- ✅ Slack routing rules are ordered most-specific-first and terminate in a `{ when: {} }` catch-all, so `rake_slack` cannot raise `NoMatchingRule`; `!cancelled()` keeps `cancelled` out of the outcome domain; `continue-on-error: true` means a Slack outage cannot fail a build or block auto-merge.
- ✅ The `skip-ci-check` → `check`/`test` → `prerelease` → `release` chain is correct: `prerelease` needs no `if:` guard, because GitHub skips a job when all its `needs` are skipped.
- ✅ `github_token` resolution fails fast with an actionable message rather than deferring to an opaque Octokit 401.
- ✅ Verified green on the real runner — the D8 native-docker watch-item is satisfied for this repo.

### General Findings

- 🟡 **Security**: **Deletion is not revocation — the removed credentials remain live.** This PR `git rm`s `config/secrets/circle_ci/config.yaml` (CircleCI API token), `config/secrets/github/config.yaml` (GitHub PAT) and `config/secrets/ci/ssh.{private,public}` (CI deploy key). All are git-crypted, so they persist as ciphertext in every prior commit of this **public** repo, decryptable by any holder of the five GPG keys under `.git-crypt/keys/default/0/`; the git-crypt symmetric key is not rotated here. Removing a credential from `HEAD` does not invalidate it. The plan's host-side sweep covers disabling the CircleCI project and deleting the `CircleCI` deploy key from repo settings, but neither the plan nor this PR revokes the **CircleCI API token** or the **GitHub PAT** at their issuer. This looks like a gap in the plan rather than an oversight in this diff, and it will replicate across all ~103 repos, where the token and PAT are plausibly shared. The PR description's "the stored token ... is deleted" reads as though the credential is gone, when only the working-tree copy is. *Not blocking this PR* — but add explicit revocation to the end-of-migration sweep, distinct from file deletion, and while there, check whether the five enrolled git-crypt GPG keys still correspond to current maintainers. That key set is the real access boundary for every historical secret in this repo.
- 🔵 **Correctness / Standards**: Conformance sweep found no deviations beyond `PLATFORMS`. §4.1–§4.8 were compared line-by-line in both directions; §4.3's key move is recorded as a 100% similarity rename; the lockfile carries `rake_slack` 0.3.0, `rake_git_crypt` 0.4.0 and `rake_github` 0.17.0 with `rake_circle_ci` and `rake_ssh` gone. The failing `CircleCI Pipeline` check on this PR is the plan's documented cosmetic "config not found" error and resolves when the project is disabled in the sweep.

---
*Review generated by /accelerator:review-pr — 5 lenses (correctness, security, safety, standards, code-quality)*

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+        if: ${{ !cancelled() }}
+        continue-on-error: true
+        run: ./go "slack:notify[${{ job.status }}]"
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+  test:
+    needs: [skip-ci-check]
+    if: needs.skip-ci-check.outputs.should_skip_ci != 'true'
+    runs-on: ubuntu-latest
```
phelma, line 58: 🔵 **[in-scope] Safety — `check`/`test` have no `timeout-minutes`, and `queue: max` turns a hang into a stalled release pipeline**

`prerelease` (line 85) and `release` (line 119) are bounded at `timeout-minutes: 30`, but `check` (line 41) and `test` (this job) carry no timeout, so they inherit GitHub's **360-minute** default. The new `concurrency: {group: main, queue: max}` serialises all `main` runs — correctly — but that means a single hung job holds the group for up to six hours, with `queue: max` faithfully queueing every subsequent merge behind it rather than superseding it.

A hang is plausible here specifically: D8 moves the integration suite onto the runner's native Docker daemon with compose-based dependency startup, which the plan itself flags as the family's headline behavioural change and an explicit watch-item (issue 06).

**Why in-scope**: the plan specifies `timeout-minutes: 30` for the publishing jobs but is silent on `check`/`test` — this is a gap the plan does not address rather than a deliberate decision, so it isn't covered by the parity principle.

**Impact**: one wedged test run stalls the repo's release pipeline for up to six hours, queue growing behind it. Worse, it is *silent* — the `Notify Slack` step only runs once the job finally ends, so nobody is told until the timeout fires. This is the 3am case: stuck and quiet.

**Suggestion**: add `timeout-minutes: 30` (matching the bound already chosen for the publishing jobs) to `check` and `test` in both workflows, so a hang releases the queue quickly and notifies. Worth applying family-wide rather than only here.

Gemfile.lock
```diff
@@ -1,18 +1,21 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (7.1.3.4)
+    activesupport (8.1.3)
```
phelma, line 4: 🔵 **[plan concern] The lockfile pins gems whose declared Ruby floor exceeds `.tool-versions` — CI passes anyway, but the plan's "don't pin back" guidance may need a floor caveat**

`activesupport (8.1.3)` declares `required_ruby_version >= 3.2.0`; so do `public_suffix (7.0.5)` and `connection_pool (3.0.2)`. `.tool-versions` still pins `ruby 3.1.1`, and `go` hard-enforces it (`ruby -v | grep -q "3.1"`). `activesupport` is not incidental — it reaches the graph via `rake_factory (0.34.0.pre.2) → activesupport (>= 4)`.

**I verified this does not break the build.** The `check` and `test` jobs on this PR both pass using `/home/runner/.asdf/installs/ruby/3.1.1/bin/ruby`; `bundle install` succeeds and the committed lockfile is what runs (the runner's `rspec-core-3.13.4` matches the lock, so there is no silent re-resolve). An earlier draft of this review called this critical; that was wrong and I've withdrawn it.

**Why plan concern**: plan §4.5 explicitly blesses this outcome — *"the targeted update will still pull in transitive major-version bumps … expected and acceptable. Do not attempt to pin transitive dependencies back"* — and §4.5 equally forbids editing `.tool-versions`. So the implementer had no plan-conformant way to avoid it, and the diff is correct as written.

**Why it's still worth a human's attention**: the plan's guidance was written for version bumps, and probably didn't anticipate one crossing the *interpreter* floor. The repo now runs gems on a Ruby their maintainers don't support — it works because nothing exercises an incompatible path, which is a latent rather than absent risk. There is no `RUBY VERSION` section in the lock, and §4.5's own `ASDF_RUBY_VERSION=3.2.11` host workaround plus the §5.4 clean-env check both run under 3.2.11, so nothing in the verification chain would ever surface this.

**Suggestion** (plan-level, fleet-wide — not this PR): decide deliberately whether the docker family moves to Ruby 3.2, or whether §4.5 should gain a "do not cross the `.tool-versions` Ruby floor" caveat. A `ruby file: '.tool-versions'` directive in the `Gemfile` would make bundler enforce the floor at resolution time and turn this class of drift into a loud error rather than a latent one.

.github/workflows/main.yaml
phelma, line 18: 🔵 **[plan concern] `queue: max` is unexercised by this PR and is a single point of total failure for `main.yaml` if the key is not accepted**

GitHub Actions rejects a workflow file **wholesale** when it contains an unrecognised key at this level — the failure mode is not "the key is ignored", it is "the workflow is invalid and never runs".

**Why plan concern**: `queue: max` is specified verbatim by plan §4.2 and cited to a GA changelog dated 2026-05-07, so the diff is conformant and the decision is documented.

**Why it's worth verifying before merge anyway**: this PR's green checks come from `pr.yaml` only — **`main.yaml` has never been parsed by Actions**, because nothing has pushed to `main`. And this same commit deletes the CircleCI pipeline, so if `main.yaml` is rejected post-merge there is no fallback: no prerelease, no release, no notification, and the failure is silent (an invalid workflow doesn't notify).

This is cheap to de-risk and the pilot repo is exactly where it should be caught, so I'd rather flag it than assume. **Suggestion**: after merge (or via a scratch branch), confirm `gh api /repos/infrablocks/docker-base-alpine-aws/actions/workflows` lists `main.yaml` as `active` rather than erroring. If the key turns out to be unsupported, dropping to plain `group: main` + `cancel-in-progress: false` degrades gracefully to "only the newest pending run is kept" — the trade-off the plan's own comment already describes.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - uses: actions/checkout@v4
+        with:
+          ref: main
+          fetch-depth: 0
+      - name: Pull latest main
+        # Approval can land long after the run starts; release publishes main
+        # as of approval time, not the tested SHA — parity with the old
+        # release.sh, which also pulled. prerelease.sh never pulled, so the
+        # prerelease job deliberately has no pull.
+        run: git pull
```
phelma, line 133: 🔵 **[plan concern] The `release` approval attests to a SHA that `git pull` then discards — and the pull may be a no-op regardless**

Two related observations on this step, merged from the security and correctness lenses.

**1. Approval and released content refer to different things.** `release` checks out `ref: main` then pulls, so what publishes is `main` as of *approval time* — not the SHA `check`/`test` validated, and not the SHA the `maintainers` reviewer saw. Anyone with push access to `main` can land a commit between approval request and approval landing, and it releases under the approver's authorisation without passing the gate.

*Why plan concern*: explicitly documented in both plan §4.2 and the PR description as parity with the old `release.sh`, which also pulled.

*Why still worth raising*: the **security property changed in migration even though the behaviour didn't**. CircleCI's `hold` was an opaque button. A GHA environment gate is presented to the reviewer as an approval *of a specific run and SHA* — so the gate now makes an attestation it does not actually enforce, and the audit trail attributes released content to an approver who never saw it. That's new, and it's a fair thing to have not noticed when writing the plan.

**2. The comment's stated mechanism doesn't match reality.** The pull likely isn't what delivers approval-time content: `environment: release` means `actions/checkout@v4` doesn't execute until *after* approval, so `ref: main` already resolves `origin/main` at that moment. The subsequent `git pull` fetches the same ref seconds later. (It's harmless — a fresh checkout can only fast-forward, so it can't hit the "who are you" failure from running before `Set CI git author`.)

**Suggestion** (post-migration, fleet-wide): drop the `git pull` and let the checkout pin the tested SHA, so approval and released content refer to the same commit. If parity must hold for now, record the resolved SHA before/after and fail if it differs from the tested SHA. Either way, the comment is worth rewording — as written it will mislead whoever debugs this pipeline next.

Rakefile
```diff
@@ -142,51 +134,70 @@ namespace :library do
...
+  end
+  if github_token.empty?
+    raise 'No GitHub token available: set GITHUB_TOKEN or run `gh auth login`'
+  end
+
+  t.access_token = github_token
+
+  # Actions store only: nothing in pr.yaml unlocks git-crypt, so dependabot
+  # runs never need the passphrase.
+  t.secrets = [
```
phelma, line 158: 🔵 **[plan concern] `File.read` of the passphrase has no locked-tree guard, unlike the credentials block 55 lines below**

Unlike `YAML.load_file`, `File.read` raises **nothing** on git-crypt ciphertext — it returns the binary blob, and `github:secrets:ensure` will happily push it to GitHub as the value of `ENCRYPTION_PASSPHRASE`. The failure is silent at provisioning time and surfaces much later as an opaque GPG import error in `prerelease`/`release`'s `Unlock git-crypt` step, a long way from the cause.

What makes this worth writing down is the **inconsistency**: `t.credentials` (line 217) defends against exactly this locked-tree condition; this read, in the same block, does not.

**Why plan concern**: §4.4 step 3 specifies this snippet verbatim, and the plan *did* consider the locked case — §7 makes an unlocked clone an explicit operator precondition (*"If it is locked, STOP and report"*). So this is a deliberate choice to handle the case procedurally rather than in code, not an oversight in the diff.

**Why still worth raising**: nothing enforces the precondition, and an operator who forgets gets no signal — just a quietly-corrupted repo secret that breaks every subsequent release. The plan's own fail-fast instinct is right there in the `github_token` block immediately above.

**Suggestion** (plan-level, fleet-wide):

```ruby
passphrase = File.read('config/secrets/ci/encryption.passphrase')
if passphrase.start_with?("\x00GITCRYPT")
  raise 'Secrets are locked: unlock git-crypt before provisioning'
end
t.secrets = [
  { name: 'ENCRYPTION_PASSPHRASE', value: passphrase.chomp }
]
```

phelma (commented):
## Code Review: #230 — Migrate pipeline from CircleCI to GitHub Actions

**Verdict:** COMMENT

Reviewed through five lenses (correctness, security, safety, standards, code quality) against the Variant B family plan. **This diff conforms to the family plan §4 with near-verbatim fidelity in both directions** — every required change is present, and the changed-file list is exactly the plan's §4 surface with nothing beyond it. The PR's own checks are green on a real runner (`check` and `test` both pass; the integration suite builds the image and runs 6 serverspec examples, 0 failures), which is the plan's §5 authoritative validation. I found no plan violations and nothing blocking. Almost everything below is a **plan concern** — a challenge to something the plan or PR description documents as deliberate — recorded so a human can revisit the *plan*, not to block this PR.

### How to read this review

Findings are labelled:

- **[in-scope]** — a genuine defect of this diff, or of something the plan does not specify. **2 findings** (1 major, 1 minor).
- **[plan concern]** — challenges a documented deliberate decision (the §1 parity list, a D-decision, or the plan's exact mandated YAML/Ruby). **10 findings.** These must not block a plan-conformant diff; they are input to post-migration hardening.

### Verified during review (false positives, not counted)

Two lens findings were **disproven by evidence** and are recorded here rather than as findings:

- 🔴→✅ **"The lockfile pins `activesupport 8.1.3` (Ruby >= 3.2) against `.tool-versions` 3.1.1, so every job fails at `bundle install`."** The Ruby-floor mismatch is real (`activesupport 8.1.3`, `public_suffix 7.0.5`, `connection_pool 3.0.2` all declare `>= 3.2`), but the predicted failure does **not** occur. Job logs confirm the runner uses `/home/runner/.asdf/installs/ruby/3.1.1/bin/ruby`, `bundle install` succeeds, and the full suite passes with the committed lockfile (`rspec-core-3.13.4` matches the lock — no silent re-resolve). The residual latent risk is recorded as a plan concern on `Gemfile.lock` instead.
- 🔵→✅ **"`Layout/ExtraSpacing` offences in the Slack routing rules."** Rubocop runs clean on the runner: *"7 files inspected, no offenses detected"*.

### Cross-Cutting Themes

- **Supply-chain mutability of CI inputs** (flagged by: security, safety) — the `docker-compose` guard fetches an unpinned `latest` binary with no checksum, and `asdf_install@v1` is a mutable tag; both run in jobs holding `contents: write` and the git-crypt passphrase. Both are plan-mandated, so this is a fleet-level hardening item.
- **Locked-tree handling is inconsistent across the Rakefile** (flagged by: correctness, code quality, safety) — `t.credentials` defends against a locked tree via `rescue Psych::SyntaxError`, but `t.secrets`'s `File.read` of the passphrase ~55 lines above has no equivalent guard. The plan assigns that case to an operator precondition (§7) rather than to code.
- **Approval is decoupled from the tested SHA** (flagged by: security, correctness) — `release` checks out `ref: main` then `git pull`s, so what ships is main at approval time. Documented parity with the old `release.sh`.

### Tradeoff Analysis

- **Parity vs hardening**: the plan's parity principle deliberately preserves inherited hazards (tag-then-publish ordering, ungated rc publishes, no update-type filter on auto-merge) so that migration risk stays separable from behaviour change. Several lenses independently rediscovered these. That is the plan working as intended — the right response is a post-migration hardening backlog, not edits to this PR.
- **Diagnosability vs plan uniformity**: narrowing the credentials `rescue` and guarding the passphrase read would both improve 3am failure modes, but both change plan-mandated code that ships to ~103 repos. Correct venue is the plan, not this repo.

### Strengths

- ✅ **Plan conformance is exemplary.** Both workflows match §4.1/§4.2 line-for-line; all six §4.4 Rakefile changes are present and correctly placed; the §4.7 decommission is complete and precisely scoped (`.circleci/`, `scripts/ci/`, the SSH pair, `circle_ci/` and `github/` configs gone; `config/secrets/dockerhub/` and the GPG key pair correctly untouched); the GPG key move is a pure 100%-similarity rename; `git grep -i circleci -- ':!Gemfile.lock'` is clean; no badge invented where none existed.
- ✅ **The security fundamentals are right**: `pull_request` (not `pull_request_target`), so fork code never sees secrets; untrusted `head_commit.message` and `pull_request.html_url` bound via `env:` rather than interpolated into `run:` — the classic script-injection vector is closed; top-level `contents: read` with `contents: write` elevated only on the two tagging jobs and `pull-requests: write` only on the merge job; `merge-pull-request` never checks out untrusted code.
- ✅ **The `dynamic` credentials change is correct end-to-end** and empirically proven: `rake_factory` re-evaluates on every parameter read, `rake_docker` guards with `Docker.authenticate!(t.credentials) if t.credentials`, and the `test` job demonstrably builds the image in a locked tree. `require 'rake_factory/kernel_extensions'` is correctly added — without it, bare `dynamic` would be a load-time `NoMethodError`.
- ✅ **`ENCRYPTION_PASSPHRASE` is Actions-store only** and `pr.yaml` never unlocks git-crypt, so dependabot runs genuinely cannot reach the passphrase — the lazy credentials load is what makes that hold, and D3 auto-merge survives.
- ✅ **`concurrency: {group: main, cancel-in-progress: false, queue: max}` is a genuine safety improvement** over CircleCI, which had no guard against close-together merges racing on version tags.
- ✅ **Fail-safe ordering throughout**: notify steps are `continue-on-error: true` so Slack outages can't fail a build or block auto-merge; the `skip-ci-check` guard propagates skips to `prerelease`/`release`, so nothing publishes in an unknown state; `fetch-depth: 0` is correctly applied to exactly the tag-resolving jobs; Slack routing rules are ordered correctly against `rake_slack`'s first-match semantics.
- ✅ **The `github_token` block fails fast with an actionable message** rather than surfacing later as an opaque Octokit 401, and the Rakefile comments consistently explain *why* rather than *what*.

### General Findings

- 🟡 **[in-scope] Security — deleted CircleCI-era credentials are removed from the tree but never revoked.** This PR `git rm`s four pieces of live credential material: the CircleCI deploy key pair (`config/secrets/ci/ssh.{private,public}`), the CircleCI API token (`config/secrets/circle_ci/config.yaml`), and the stored GitHub PAT (`config/secrets/github/config.yaml`). **Deletion is not revocation.** All three credentials remain valid at their providers, and the Rakefile entries that managed them are now gone, so nothing in the repo tracks, rotates, or can even enumerate them — they become orphaned standing access. They also persist in git history; they are git-crypted, so their confidentiality now rests on the CI GPG key this same PR relocates to a public repo path.

  Deleting the deploy key from repo settings is a **documented** deferral (plan §1 out-of-scope, PR description). Revoking the **CircleCI API token and the GitHub PAT** does not appear to be tracked anywhere — which is why I count this as in-scope rather than a plan concern. D7's own reasoning (a long-lived stored token is the wrong shape) applies with equal force to leaving it alive but untracked.

  *Suggestion*: revoke the GitHub PAT and the CircleCI API token at their providers when this merges, and track each with an owner and deadline rather than relying on the end-of-migration sweep to remember. Purging history is not warranted given git-crypt protection — revocation is the control that matters.

---
*Review generated by /accelerator:review-pr — 5 lenses (correctness, security, safety, standards, code quality)*

Rakefile
```diff
@@ -203,9 +214,11 @@ namespace :image do
     t.repository_name = 'alpine-aws'
     t.repository_url = 'infrablocks/alpine-aws'
 
-    t.credentials = YAML.load_file(
-      'config/secrets/dockerhub/credentials.yaml'
-    )
+    t.credentials = dynamic do
```
phelma, line 217: 🔵 **[plan concern] The credentials `rescue` conflates "tree is locked" with "YAML is malformed"** *(flagged independently by correctness, code quality, and safety)*

The lazy load itself is **correct and verified** — I confirmed `rake_factory` re-evaluates the `dynamic` value on every parameter read and `rake_docker` guards with `Docker.authenticate!(t.credentials) if t.credentials`, and the `test` job on this PR demonstrably builds the image in a locked tree. Only the *breadth* of the rescue is in question.

`rescue Psych::SyntaxError → nil` treats two different causes as one:

1. **Locked tree** — nil credentials are correct and intended (base images are public). This is the case the plan is solving for.
2. **A genuine typo in the decrypted `credentials.yaml`** — in the *unlocked* `prerelease`/`release` jobs, this also yields nil, so `image:publish` skips `Docker.authenticate!` and the run dies at push time with an opaque Docker Hub 401, swallowing the "your YAML is broken" message the exception was carrying. And because `version:bump[rc]` has *already pushed the tag* by then, a clean pre-tag abort becomes an instance of the documented dangling-tag hazard.

(The safety lens checked whether this could silently publish an unauthenticated image — it can't: `RakeDocker::Output.print` raises on registry error chunks, so the push fails loudly. The cost is diagnosability, not a false green.)

**Why plan concern**: §4.4 step 2a specifies this exact snippet, and it is called out as the one deliberate exception to leaving the `image` namespace untouched (D5).

**Suggestion** (plan-level — it ships to ~103 repos): key the decision off lock *state* rather than parse failure, returning nil only for git-crypt ciphertext and letting everything else propagate:

```ruby
t.credentials = dynamic do
  path = 'config/secrets/dockerhub/credentials.yaml'
  next nil if File.binread(path, 9) == "\x00GITCRYPT"
  YAML.load_file(path)
end
```

This also closes a smaller edge the correctness lens raised: git-crypt ciphertext is arbitrary binary, and depending on where invalid bytes land, `YAML.load_file` can surface `ArgumentError: invalid byte sequence in UTF-8` rather than `Psych::SyntaxError` — which the current rescue would not catch. It works today, but it rests on the error shape happening to match.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install tools
+        uses: infrablocks/github-actions/asdf_install@v1
+      - name: Ensure docker-compose is available
```
phelma, line 32: 🔵 **[plan concern] The docker-compose guard fetches an unpinned `latest` binary as root with no integrity check** *(flagged by security and safety; duplicated in `main.yaml`)*

This step downloads `docker-compose` from the `/releases/latest/download/` redirect, `sudo chmod +x`es it into `/usr/local/bin`, and executes it — no version pin, no checksum, no signature. In `main.yaml` the same binary runs in the same workflow as jobs holding `contents: write` and `ENCRYPTION_PASSPHRASE`.

**Why plan concern**: this reproduces plan §4.1/§4.2's mandated snippet verbatim — the diff is fully conformant. Raising it because the plan doesn't appear to have weighed the supply-chain property, not because the implementer deviated.

**Why it has teeth at fleet scale**: `latest` resolves at run time, so an upstream Compose regression breaks the integration tests in *every* migrated Docker repo simultaneously, with no pinned version to roll back to — and the fix is a fleet-wide edit under time pressure. The step is also **dormant today** (the runner ships the `docker compose` plugin, so the `command -v docker-compose` guard skips the download) and would activate silently on a future runner-image change. Correlated *and* surprising.

**Suggestion** (plan-level, post-migration): pin an explicit version and verify a checksum —

```bash
DOCKER_COMPOSE_VERSION=v2.x.y
sudo curl -fsSL ".../download/${DOCKER_COMPOSE_VERSION}/docker-compose-linux-x86_64" -o /usr/local/bin/docker-compose
echo "<sha256>  /usr/local/bin/docker-compose" | sha256sum -c -
sudo chmod +x /usr/local/bin/docker-compose
```

Alternatively, shim the runner's preinstalled `docker compose` plugin and avoid the download entirely.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Install tools
+        uses: infrablocks/github-actions/asdf_install@v1
```
phelma, line 135: 🔵 **[plan concern] `asdf_install@v1` is a mutable reference running in the jobs that hold write access and the passphrase**

In `prerelease`/`release`, this action runs after a checkout that persists a `contents: write` `GITHUB_TOKEN` into `.git/config`, and in the same job as the `ENCRYPTION_PASSPHRASE` unlock. A tag is mutable: anyone who can move `v1` in the actions repo — or who compromises it — gains code execution in every InfraBlocks release job simultaneously, with no per-repo change to detect it.

**Why plan concern**: D6 and the PR description document this explicitly (*"our own action; we are happy tracking its major version tag"*).

**Why still worth noting**: the **trust decision is reasonable** and I'm not disputing it — the concern is the *mutability of the reference*, not the publisher. Those are separable, and you can keep the former while fixing the latter. Blast radius is ~103 repos at once.

**Suggestion** (fleet-wide, post-migration): keep the same trust boundary but make the reference immutable — `uses: infrablocks/github-actions/asdf_install@<sha> # v1`, with Dependabot's `github-actions` ecosystem bumping it (this repo already has Dependabot wired up). Protecting the `v1` tag in the actions repo so it can't be moved without review is a lighter-weight alternative.

README.md
```diff
@@ -2,16 +2,16 @@
...
-To encrypt a GPG key for use by CircleCI:
+To encrypt a GPG key for use by CI:
 
 ```bash
 openssl aes-256-cbc \
   -e \
   -md sha1 \
   -in ./config/secrets/ci/gpg.private \
-  -out ./.circleci/gpg.private.enc \
+  -out ./.github/gpg.private.enc \
```
phelma, line 14: 🔵 **[plan concern] The CI GPG key is publicly downloadable, protected only by a legacy openssl KDF**

This PR moves the openssl-encrypted CI GPG private key to `.github/gpg.private.enc` in a **public** repository, and this README snippet documents producing it with `openssl aes-256-cbc -md sha1` — no `-pbkdf2`, no `-iter`. That invokes the legacy `EVP_BytesToKey` KDF: a single SHA-1 pass over the passphrase plus an 8-byte salt, i.e. essentially no work factor against offline cracking. The only barrier between an anonymous internet user and the CI GPG key — and therefore the entire git-crypt tree, including the Docker Hub credentials and the deleted-but-still-valid SSH deploy key in history — is the entropy of `ENCRYPTION_PASSPHRASE`.

**Why plan concern**: §4.6 explicitly says to leave the openssl commands untouched, and the key was equally exposed at `.circleci/gpg.private.enc` before this PR. **This is inherited, not introduced** — the diff changes only the path.

**Why raise it here**: the migration is the natural moment to fix it (the key is being touched and re-documented anyway), and the exposure is fleet-wide. Note this also interacts with the unrevoked-credentials finding in the summary: breaking this passphrase recovers a *still-valid* repo-write deploy key from git history.

**Suggestion** (post-migration, fleet-wide): re-encrypt with a modern KDF — `openssl aes-256-cbc -pbkdf2 -iter 600000 -md sha256` — updating both README snippets and confirming `rake_git_crypt`'s unlock passes matching flags. As an interim control, confirm `encryption.passphrase` is high-entropy and machine-generated (the `encryption:passphrase:generate` task suggests it is, which would make this a low-likelihood/high-impact item rather than an urgent one).

Rakefile
```diff
@@ -142,51 +134,70 @@ namespace :library do
...
-  github_config =
-    YAML.load_file('config/secrets/github/config.yaml')
-
-  t.access_token = github_config['github_personal_access_token']
-  t.deploy_keys = [
-    {
-      title: 'CircleCI',
-      public_key: File.read('config/secrets/ci/ssh.public')
-    }
+  # Operator's ambient auth — the stored PAT is gone (see the cutover PR's
```
phelma, line 141: 🔵 **[plan concern] The comment points at an unnamed PR that will be hard to find later**

"see the cutover PR's deliberate-decisions list" is an unnumbered, unlinked reference. Once this merges and the migration sweep finishes, a maintainer wondering why provisioning uses ambient `gh` auth has to go archaeology-hunting through PR history to find the rationale — and this is precisely the moment someone is tempted to "helpfully" reinstate a stored PAT.

**Why plan concern**: §4.4 step 3 mandates this comment text verbatim. The rest of it is genuinely good — it explains the fail-fast intent well, and the `Errno::ENOENT` rescue correctly covers gh-not-installed. Only the dangling reference is weak.

**Suggestion** (plan-level, so all ~103 repos inherit it): make the pointer durable — cite the PR number inline (`see PR #230`), or better, reference the org decision record (D7) that this derives from, so the link survives the PR falling off the first page. Low priority; noted for the plan's next revision rather than for this PR.

phelma:
Superseded: the family plan gained amendments from the fleet review (git-crypt header check, release-gate concurrency design, toolchain drift fix, pinned compose, timeouts). Re-running the migration pipeline from the amended plan on a fresh branch.

PR #232, opened 2026-07-20

phelma:

  inside the `image` namespace

  `lib/version.rb` logic.
  Updates ship with the next human-triggered release.
  improvement.
- The `release-gate` job asserts the `release` environment has protection
  rules before approving — GitHub silently auto-creates an unprotected
  environment if provisioning never ran, which would release ungated.
  supplies secrets/context.
  hold a stable lockfile on CI.

end-of-migration sweep.


phelma:
Closing: prerelease publish failed due to a bug in the migration task template (docker-api push resolved the repository from the bare local tag). Template fixed; superseding PR to follow from a fresh pipeline run.

PR #233, opened 2026-07-20

phelma:

  inside the `image` namespace

  `lib/version.rb` logic.
  Updates ship with the next human-triggered release.
  improvement.
- The `release-gate` job asserts the `release` environment has protection
  rules before approving — GitHub silently auto-creates an unprotected
  environment if provisioning never ran, which would release ungated.
  supplies secrets/context.
  hold a stable lockfile on CI.

end-of-migration sweep.

---

🏭 This PR was opened by **Foundry**, Atomic's AI software development
factory. Implementation, review, and fixes are performed by AI agents;
merges happen automatically once the review and checks gates pass.
This task migrates a Docker image repo's CI from CircleCI to GitHub Actions.

| Pipeline | Task pack | Family | Run |
|---|---|---|---|
| `migrate` | `migrate-docker` | docker_image | `2026-07-20T11-02-34-015Z` |

<sub>atomic-foundry-pr · foundry-pipeline: migrate · foundry-task: migrate-docker · foundry-run: 2026-07-20T11-02-34-015Z</sub>


phelma:
### Plan-conformance: `rake_factory` resolves to a pre-release version

`Gemfile.lock` pins `rake_factory (0.34.0.pre.2)`. A stable `0.33.0` exists on rubygems.org (`gem list -r -a rake_factory` → `0.33.0, 0.31.0, …`; there is no stable 0.34.0), and every dependant in the lock requires only `~> 0.33`.

Plan §4.5 asks for a newly-declared `gem 'rake_factory'` at `>= 0.32` — it does not ask for a pre-release. Shipping a `.pre` gem into the production lockfile means CI (and `prerelease:publish`/`image:publish`, which depend on `dynamic` from this gem) runs against an unreleased build that can be yanked or changed.

**Suggested fix**: constrain to the stable line and re-lock, e.g. `bundle lock --update=rake_factory` with `gem 'rake_factory', '~> 0.33'` in the Gemfile, then re-run the §5.4 clean-env install to confirm.

---

### Note (plan concern, not a diff defect): §5.8 grep check cannot pass as specified

Plan §5.8 requires `git grep -i circleci -- ':!Gemfile.lock'` to match nothing, but §4.7 mandates the body of `scripts/ci/common/assert-release-environment-protected.sh` verbatim — and that body contains the word "CircleCI" in its comment. The two plan clauses contradict each other, and the same comment also sits awkwardly against §4's comment rule ("never reference … deleted CircleCI scripts").

The implementation followed §4.7 verbatim, which is the right call. Flagging so the family plan can be reconciled — either exempt that file from the grep, or reword the comment to drop the CircleCI reference.

phelma (commented):
## Code Review: #233 - Migrate pipeline from CircleCI to GitHub Actions

**Verdict:** COMMENT

Five lenses (correctness, security, safety, standards, code quality) reviewed this cutover against the Variant B family plan. The diff is a near-verbatim implementation of the plan's §4 templates — both workflows, the Rakefile changes, the decommission, and the lockfile/platform hygiene all conform, and the standards lens found essentially no drift. The great majority of lens findings challenge decisions the plan or PR description documents as deliberate; those are recorded below as **plan concerns** so a human can revisit the family plan, but they do not block a plan-conformant diff. One genuine in-scope defect was found (a pre-release `rake_factory` in the lockfile), and one lens finding was disproven by evidence.

### Cross-Cutting Themes

- **`queue: max` concurrency key** (flagged by: correctness, safety) — both lenses believe this is not a valid GitHub Actions `concurrency` key and that the documented "keep every queued run" guarantee therefore does not hold. **Plan concern**: §4.2 specifies this key verbatim and cites the 2026-05-07 GA changelog for larger queues. The implementation is correct against the spec; if the lenses are right, the plan needs correcting fleet-wide, and the first real `main` run will settle it.
- **`docker-compose` pin `v5.3.1` + checksum** (flagged by: correctness, safety) — both lenses believe this release tag does not exist (compose is on the 2.x line). **Plan concern**: §4.7 mandates this exact version and sha256. If the pin is wrong it is wrong in every Variant B repo, so it should be fixed in the plan, not here. Worth confirming against the release assets before the first `test` job runs on a runner lacking a preinstalled `docker-compose`.
- **Release publishes main-at-approval-time, not the approved SHA** (flagged by: correctness, safety) — `ref: main` plus `git pull --ff-only` means a commit landing during the approval window ships unapproved. **Plan concern**: §4.2's exact YAML and its inline comment specify precisely this behaviour, mirroring `release.sh`'s `git pull`.
- **Docker Hub credentials degrade to `nil` on a locked tree** (flagged by: security, safety) — both would prefer the push paths fail loudly rather than proceed anonymously. **Plan concern**: §4.4 step 2a reasons about this explicitly and rejects the fail-loud alternative (a rescue also fires on genuinely malformed credentials in an unlocked tree).

### Tradeoff Analysis

- **Supply-chain hardening vs family uniformity**: security wants `actions/checkout@v4` and `infrablocks/github-actions/asdf_install@v1` pinned to commit SHAs, since these jobs hold the git-crypt passphrase and registry credentials. The plan (§1, D6) explicitly accepts tracking the moving `v1` tag for our own action, and §4's YAML uses `@v4`. Recommendation: keep as-is for cutover parity; raise SHA-pinning as post-migration fleet-wide hardening, where it belongs.
- **DRY vs plan fidelity**: code quality flags the git-crypt ciphertext sniff appearing three times and `prerelease:publish` reimplementing the publish path. Both are the plan's verbatim code, and the plan explains why `prerelease:publish` does its own single-tag push (`image:publish` would also overwrite `latest`). Recommendation: if the duplication is worth removing, do it once across the family, not in this repo.

### Strengths

- ✅ Near-perfect conformance to the plan's templates — workflows, Rakefile §4.4 steps 1–7, and §4.7 decommission all match, with parameterisation correctly limited to `alpine-aws` / `infrablocks/alpine-aws`
- ✅ No `${{ }}` interpolation of attacker-controlled data into `run:` — PR title, number, URL and head SHA all pass through `env:`
- ✅ Least privilege: top-level `contents: read`, `contents: write` only on the main tag-pushing jobs, `pull-requests: write` only on the dependabot merge job, and `contents: read` on the PR prerelease job
- ✅ PR prerelease and dependabot merge both gate on `pull_request.user.login` (the immutable author) rather than `github.actor`, closing the `@dependabot rebase` widening path
- ✅ `--match-head-commit "$HEAD_SHA"` makes the auto-merge fail rather than merge an unchecked head
- ✅ `assert-release-environment-protected.sh` converts GitHub's fail-open auto-created environment into a hard failure — real defence in depth
- ✅ git-crypt ciphertext detection prevents `github:secrets:ensure` silently uploading ciphertext as the passphrase
- ✅ `fetch-depth: 0` correctly applied to exactly the tag-reading jobs; `check` stays shallow
- ✅ Explicit `timeout-minutes` on every job; `cancel-in-progress: false` on both tag-pushing jobs
- ✅ Comment rule honoured — comments are repo-local and self-contained, with no reference to the plan, D-numbers, or the migration process
- ✅ Lockfile hygiene: `PLATFORMS` reduced to `ruby` + `x86_64-linux`, no operator-local darwin entries, `rake_circle_ci`/`rake_ssh` fully gone, `rake_git_crypt 0.4.0` / `rake_github 0.17.0` present
- ✅ GPG key move recorded as a pure rename; `config/secrets/ci/` correctly retains only `encryption.passphrase`, `gpg.private`, `gpg.public`
- ✅ Ruby bumped to 3.3.11 per the §4.5 policy; all three retained scripts are mode 100755

### In-Scope Findings

- 🟡 **Plan conformance**: `Gemfile.lock` pins `rake_factory (0.34.0.pre.2)`, a pre-release, where stable `0.33.0` exists and every dependant requires only `~> 0.33`. §4.5 asks for `>= 0.32`, not a pre-release. Posted as a separate PR comment with a suggested fix.
- 🔵 **Standards / Code Quality**: `scripts/ci/common/install-git-crypt.sh` retains the CircleCI-era preamble (`[ -n "$DEBUG" ] && set -x`, `set -e`, `set -o pipefail`) while the two sibling scripts use `set -euo pipefail`. The plan only specifies the body, so this is conformant — but `scripts/ci/` now holds three scripts in two styles, and the surviving `DEBUG` idiom is a leftover from a CI system this repo no longer uses.
- 🔵 **Code Quality**: in `install-docker-compose.sh` the version (`v5.3.1`, inside the URL) and its checksum are separated bare literals with no named binding. Hoisting to `COMPOSE_VERSION` / `COMPOSE_SHA256` would make a future bump harder to get half-right. (Plan-specified body, so a plan-level suggestion.)
- 🔵 **Safety (suggestion)**: no documented rollback path for a bad release — the pipeline pushes both git tags and Docker Hub tags, both effectively irreversible for consumers, and recovery depends on tribal knowledge. A short README section would help; arguably fleet-wide work.

### Plan Concerns (documented deliberate decisions — do not block)

Recorded for a human to revisit at plan level, not defects of this diff:

- 🔴 `queue: max` may not be a valid concurrency key (correctness, safety) — §4.2 verbatim, with GA citation
- 🔴 `docker-compose v5.3.1` pin may not exist (correctness, safety) — §4.7 verbatim
- 🟡 Actions referenced by mutable tag on secret-bearing jobs (security) — §1/D6 explicitly accepts this
- 🟡 PR prerelease publishes to the production Docker Hub repo for any same-repo collaborator PR (security) — D9 / §4.4 step 7, explicitly in scope and reasoned; accumulating PR tags is a documented accepted cost
- 🟡 Release publishes main-at-approval-time rather than the approved SHA (correctness, safety) — §4.2 verbatim, mirrors `release.sh`
- 🟡 `Docker::Image.all(filter: 'alpine-aws').first` is a non-deterministic image lookup (correctness) — §4.4 step 7 verbatim
- 🟡 `prerelease:publish` reimplements the publish path rather than reusing `image:publish` (code quality) — §4.4 step 7 explains why (`image:publish` would also overwrite `latest`)
- 🟡 git-crypt ciphertext sniff duplicated three times in the Rakefile (code quality) — all three sites are plan-verbatim
- 🔵 Docker push stream errors are not inspected, so a failed push can report success (correctness) — §4.4 step 7 verbatim
- 🔵 Credentials degrade to `nil`/anonymous on a locked tree instead of failing loudly (security, safety) — §4.4 step 2a explicitly rejects the fail-loud alternative
- 🔵 `assert-release-environment-protected.sh` counts any protection rule rather than asserting `required_reviewers` specifically (correctness) — §4.7 verbatim
- 🔵 `install-docker-compose.sh`'s `command -v` short-circuit means the pin and checksum are skipped whenever the runner already ships a `docker-compose` (security) — §4.7 specifies "installing only if missing"
- 🔵 `.github/gpg.private.enc` uses OpenSSL's legacy `-md sha1` KDF with no PBKDF2 (security) — pre-existing; §4.6 says leave the openssl commands untouched
- 🔵 `github_token` resolution runs at Rakefile load time for every invocation, including unrelated tasks (code quality) — §4.4 step 3 verbatim
- 🔵 Slack channel IDs are opaque literals with names only in trailing comments (code quality) — §4.4 step 4 verbatim
- 🔵 §5.8's `git grep -i circleci` check cannot pass, because §4.7's mandated script body contains "CircleCI" in a comment — the plan contradicts itself; the implementation correctly followed §4.7. Posted as a PR comment.

### Disproven (false positive)

- The correctness lens flagged `latest_tag.rc!` as an in-place mutation that could contaminate later readers. Verified against `lib/version.rb:16-30`: `rc!` delegates to `start_rc`/`increment_rc`, both of which `clone` before mutating and return a new object. `latest_tag` is not mutated. No issue.

---

phelma (commented):
## Code Review: #233 - Migrate pipeline from CircleCI to GitHub Actions

**Verdict:** COMMENT

This cutover is a faithful, high-fidelity rendering of the Variant B family plan: both workflows match the §4.1/§4.2 templates essentially character-for-character, the Rakefile edits (lazy git-crypt-guarded Docker Hub credentials, `rake_github` provisioning, Slack routing, `repository:set_ci_author`, `pipeline:prepare`, the additive `prerelease:publish` task) are all present and correctly parameterised to `alpine-aws` / `infrablocks/alpine-aws`, and the decommission is complete. No critical or blocking defects were found across the correctness, security, safety, standards and code-quality lenses — the remaining items are minor robustness/consistency polish plus a set of plan-documented deliberate decisions recorded here for a future human to revisit.

### Cross-Cutting Themes

- **`Docker::Image.all(filter: 'alpine-aws').first` is an implicit, order-dependent image selection** (flagged by: Correctness, Code Quality) — correct on an ephemeral `ubuntu-latest` runner where only one match exists, but a substring filter plus `.first` would silently tag and push the wrong image on a warm or self-hosted daemon, and the `ImageNotFound` message names neither what was searched for nor what was found.
- **`scripts/ci/common/install-git-crypt.sh` retains the CircleCI-era shell preamble** (flagged by: Standards, Code Quality) — `[ -n "$DEBUG" ] && set -x` / `set -e` / `set -o pipefail`, while its two sibling retained scripts both use `set -euo pipefail`. Three surviving scripts, two conventions; this one also lacks `-u`.
- **Release publishes `main` as of approval time, not the approved SHA** (flagged by: Security, Safety) — plan-mandated (§4.2 mirrors `release.sh`'s `git pull`), noted as a time-of-check/time-of-use gap in the release authorisation for the post-migration hardening list.

### Tradeoff Analysis

- **Parity vs. safety on the release checkout**: the `ref: main` + `git pull --ff-only` shape reproduces CircleCI exactly, at the cost of the environment approval not binding to a commit. Correct call at cutover; pinning `ref: ${{ github.sha }}` is the eventual fix, fleet-wide.
- **Fail-open Slack vs. observability**: `continue-on-error: true` on every notify step correctly prevents a Slack outage failing a build or blocking dependabot auto-merge, at the cost of a silent failure if Slack is down during a release failure. Right trade (D2), worth knowing about.

### Strengths

- ✅ No script injection: PR title, URL, number and head SHA all flow through `env:` and are referenced as quoted shell variables — the classic `${{ }}`-into-`run:` vector is closed throughout.
- ✅ Trigger is `pull_request`, not `pull_request_target`; top-level `permissions: contents: read` with narrow per-job elevation (`contents: write` only on the tag-pushing jobs, explicit `contents: read` on the PR prerelease job).
- ✅ Dependabot auto-merge keys on the immutable `github.event.pull_request.user.login` and pins with `--match-head-commit "$HEAD_SHA"`, closing the post-checks-push window.
- ✅ `merge-pull-request` correctly keeps `needs: [check, test]` only, avoiding the cascade-skip that adding `prerelease` would cause.
- ✅ The `\0GITCRYPT` magic-header guards fail loudly on genuinely malformed credentials in an unlocked tree, rather than the `rescue Psych::SyntaxError` shape that would silently publish with nil auth after the version tag was pushed.
- ✅ `assert-release-environment-protected.sh` converts GitHub's fail-open environment behaviour into fail-closed, with an actionable message naming the fix.
- ✅ Gate/slot concurrency split is right: `main-tags` with `cancel-in-progress: false` + `queue: max` on the publishing jobs only, approval on the slot-free `release-gate`, so a pending approval never freezes main CI nor kills a mid-flight publish.
- ✅ `install-docker-compose.sh` pins an exact release and verifies sha256 in a scratch path before `sudo install`.
- ✅ `timeout-minutes` on every job (30/10) replaces GitHub's 360-minute default.
- ✅ Comments explain the non-obvious *why* (`repo_tag` on push, `run_attempt` in the PR tag, job-level concurrency) and respect the plan's comment rule — self-contained, no references to the migration, plan or deleted scripts.
- ✅ Plan conformance verified independently: `.tool-versions` at 3.3.11; Gemfile alphabetical with `rake_circle_ci`/`rake_ssh` gone and `rake_factory`/`rake_slack` added; lock carries `rake_git_crypt 0.4.0`, `rake_github 0.17.0`, `rake_slack 0.3.0`; `scripts/ci/` retains exactly the three expected executable scripts; `config/secrets/ci/` retains only `encryption.passphrase`/`gpg.private`/`gpg.public`; `circle_ci/` and `github/` gone, `dockerhub/` untouched; GPG key moved to `.github/`.

### General Findings

- 🔵 **Correctness**: `prerelease:publish` calls `latest_tag.rc!` where `latest_tag` is `nil` for a tagless repo (or an inadvertently shallow checkout), raising an opaque `NoMethodError` instead of naming the cause. A one-line guard would make the failure comprehensible.
- 🔵 **Correctness / Code Quality**: image selection via substring filter + `.first` (see cross-cutting theme above); consider selecting by exact RepoTag and enriching the `ImageNotFound` message.
- 🔵 **Correctness**: the release job's bare `git pull --ff-only` works only because `actions/checkout` creates a tracking branch when `ref` names a branch; `git pull --ff-only origin main` would be explicit and robust to a future `ref` change.
- 🔵 **Code Quality**: the `File.binread(...).start_with?("\0GITCRYPT")` idiom is now duplicated at three sites in the Rakefile. A single `git_crypt_locked?(path)` helper would make each call site read as intent — behaviour-identical, purely a DRY/naming change. (Note this is plan-prescribed shape, so it is a fleet-wide question rather than a repo-local one.)
- 🔵 **Code Quality / Standards**: normalise `install-git-crypt.sh` to `set -euo pipefail` to match its two siblings.
- 🔵 **Standards**: the `image.nil?` guard is collapsed to a trailing modifier-`if` where the plan's template uses a multi-line block. Fits within 80 chars here; flagged only because it makes fleet-wide diffing of migrated Rakefiles marginally noisier.
- 🔵 **Code Quality**: `prerelease:publish` is the one piece of genuinely new logic with no test seam — the tag grammar `"#{latest_tag.rc!}.pr#{pr}.#{run}.#{attempt}"` is only exercisable via a live PR build. Extracting the pure tag-construction into an assertable method is the cheap version of a fix; not worth restructuring for this migration.
- 🔵 **Safety**: `prerelease:publish` does not validate that all three of `pr_number`/`run_number`/`run_attempt` are present; a mis-invocation would produce a truncated tag like `…-rc.1.pr..`. Blast radius is confined to the PR tag space (disjoint from real releases by construction), so optional hardening only.

### Plan Concerns (documented deliberate decisions — not defects of this diff)

Recorded so a human can revisit the plan; none should block this PR.

- 🔵 **Security**: the PR `prerelease` job (D9) exposes `ENCRYPTION_PASSPHRASE` and Docker Hub credentials to unreviewed branch code. The `if` correctly excludes forks and dependabot, so exposure is bounded to accounts that already hold write access — but with no branch protection on these repos, "write access" and "secret access" collapse into one privilege level. Post-migration hardening candidate: an `environment:` gate on the passphrase, or OIDC / short-lived registry tokens.
- 🔵 **Security / Safety**: `release` publishes `main` at approval time rather than the approved SHA (see tradeoff above).
- 🟡 **Safety**: `git rm` of `config/secrets/{circle_ci,github}/config.yaml` and `config/secrets/ci/ssh.*` does not revoke anything — the CircleCI API token, the stored GitHub PAT and the deploy key all remain live and recoverable (git-crypted) from this public repo's history, and the `CircleCI` deploy key is still registered in repo settings with nothing managing it. Explicitly host-side out-of-scope work (issue 20) and noted in the PR description; flagged only to confirm all three items are on that sweep.
- 🔵 **Safety**: the inherited irreversible-operation hazards are preserved by design — tag-then-publish ordering (a failed publish leaves an orphan version tag), ungated rc publishes on every push to `main`, and permanent accumulation of PR prerelease tags on Docker Hub (D9 "Accepted cost").
- 🔵 **Safety**: with no concurrency group on `release-gate`, several runs can sit at the gate and be approved out of order; each `release` then pulls the then-current `main`, so later approvals may re-release an identical tree. Consistent with the plan's gate/slot design; operational mitigation is to approve only the newest queued run.
- 🔵 **Standards**: `scripts/ci/common/assert-release-environment-protected.sh` is reproduced verbatim from the plan's §4.7 template, whose comment body mentions CircleCI. That is the sole remaining `git grep -i circleci` hit, so §5 check 8 ("matches nothing") cannot be satisfied while §4.7's template is followed literally — a self-contradiction in the plan, not a defect here. Worth resolving fleet-wide, since §4's comment rule also asks that repo comments not reference the migration.

### False Positives (checked and cleared)

- `${{ job.status }}` interpolated into the notify steps' `run:` body is **not** a script-injection vector: it is a GitHub-generated enum (`success`/`failure`/`cancelled`), not attacker-controlled.
- `assert-release-environment-protected.sh` was checked for a fail-open hole: under `set -euo pipefail`, a non-2xx `gh api` response aborts rather than yielding an empty `count`, and a null `.protection_rules` yields `0` — every plausible failure mode denies the release.

---

phelma:
Closing to re-run this migration fresh through the updated pipeline.

PR #237, opened 2026-07-30
Migrate pipeline from CircleCI to GitHub Actions PP-710

phelma:
Part of PP-710.

  inside the `image` namespace

  `lib/version.rb` logic.
  Updates ship with the next human-triggered release. The job checks out and
  installs tools solely so it can carry the same `Notify Slack` step as every
  other job (CircleCI's `merge_pull_request` ended in `- notify`); on
  dependabot-triggered runs that token resolves from the Dependabot secret
  store, so it stays inert until that store is provisioned org-wide.
  improvement.
- The `release-gate` job asserts the `release` environment carries a
  required-reviewers rule before approving — GitHub silently auto-creates an
  unprotected environment if provisioning never ran, which would release
  ungated.
- `main.yaml`'s `prerelease`/`release` jobs set `queue: max` under
  `concurrency`. This is a supported key alongside `group` and
  `cancel-in-progress`
  ([docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency),
  [GA 2026-05-07](https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/)).
  It keeps every queued run rather than the default of retaining only the
  newest pending one, so close-together merges serialise FIFO on tag
  operations instead of cancelling each other.
  supplies secrets/context.
  hold a stable lockfile on CI.

end-of-migration sweep.

---

🏭 This PR was opened by [**Foundry**](https://github.com/atomicinnovation/foundry), Atomic's AI software development
factory. Implementation, review, and fixes are performed by AI agents;
merges happen automatically once the review and checks gates pass.
This task migrates a Docker image repo's CI from CircleCI to GitHub Actions.

| Pipeline | Task pack | Family | Run |
|---|---|---|---|
| `migrate` | `migrate-docker` | docker_image | `2026-07-30T15-13-42-256Z` |

<sub>atomic-foundry-pr · foundry-pipeline: migrate · foundry-task: migrate-docker · foundry-run: 2026-07-30T15-13-42-256Z</sub>


phelma:
Status from the automation that opened this pull request — updated as it progresses.

### Assessment

I checked this repository against the plan we use for migrating InfraBlocks Docker image repos from CircleCI to GitHub Actions, and it is a straightforward fit. It builds and publishes the 'alpine-aws' image to Docker Hub, versions itself with git tags, keeps its secrets under git-crypt, and its CircleCI setup follows exactly the test / prerelease / approve / release shape the plan expects, so the standard migration can be applied without improvisation. Two things need attention while migrating: the repo still pins Ruby 3.1.1, which our policy says must move to 3.3.11, and its dependency lockfile pins a version of the nokogiri library that has since been withdrawn, so the lockfile will need refreshing at the same time or CI will fail to install dependencies. It has no docker-compose test dependencies and no CI badge in its README, both of which the plan allows for. I also found two small disagreements with the offline survey: this checkout's secrets are unlocked rather than locked, and the survey listed no deviations when there are in fact several benign ones.

### Changes

I migrated this repository's CI from CircleCI to GitHub Actions. There are now two workflows: one that runs on every pull request (code checks, the integration test suite, a pre-release image publish for same-repo human PRs, and automatic merging of Dependabot updates), and one that runs on pushes to main (checks, tests, a pre-release publish, then a manually approved release). The CircleCI configuration, its step scripts and the credentials only it used have been removed, and the encrypted CI GPG key has moved from .circleci/ to .github/. The Rakefile now provisions its own GitHub secrets and the approval-gated "release" environment instead of configuring CircleCI, and sends build notifications to Slack. Along the way I bumped Ruby from 3.1.1 to 3.3.11 to match our org baseline and refreshed the dependency lockfile so it installs cleanly on a Linux CI runner. The build, publish and versioning logic itself is unchanged. I ran the code checks and the integration suite before and after the change — both were green both times — and confirmed the new GitHub secrets and release approval gate are in place.

### Verification

I checked the GitHub configuration for infrablocks/docker-base-alpine-aws to confirm the repository is ready for its new GitHub Actions pipeline. The ENCRYPTION_PASSPHRASE secret that CI needs is present in the repository's Actions secrets (I only listed secret names — no values were read). A 'release' environment exists, and it is protected by a required-reviewers rule pointing at the infrablocks Maintainers team, so releases cannot run without human approval. Everything the setup step was supposed to create is in place.

### Automated checks

Automated checks passed first time.

### Smoke test

I confirmed that this pull request's CI run actually published a working pre-release image to Docker Hub. The run (number 12, attempt 1) published the image under the tag 0.17.0-rc.11.pr237.12.1 in the infrablocks/alpine-aws repository, and I found exactly that one tag when I asked the Docker registry directly for its tag list. I then downloaded the image successfully, which proves it really exists in the registry rather than just being listed somewhere. Finally I started a throwaway container from it and asked the AWS CLI inside for its version: it reported aws-cli/1.18.124, which matches the version the image's Dockerfile pins. So the image builds, publishes, downloads and runs with the expected tooling present. I removed the downloaded image afterwards to leave the machine as I found it.

### Automated review

I reviewed this CircleCI-to-GitHub-Actions migration through five lenses — correctness, security, safety, standards and code quality — and checked it line by line against the migration specification it was written to satisfy. It matches that specification exactly: the two new workflow files reproduce the prescribed shape verbatim, the Rakefile changes are all present and correct, the old CircleCI pipeline is fully removed with the three still-needed scripts retained, the encrypted CI key move is byte-identical, and the Ruby version bump comes with a clean, settled dependency lockfile. I found nothing that would break the pipeline or leak a secret. Almost every point the review raises argues with a decision the specification and the PR description already document as deliberate — pre-merge image publishing from PR builds, releasing the branch as of approval time, the tag-before-publish ordering — so I have recorded those for a later fleet-wide conversation rather than treating them as problems with this change. The single genuinely new suggestion is small: the newly added Slack gem is the only one added here without a version pin, and pinning it would guard against a future dependency update quietly changing the notification behaviour. I consider this safe to merge.

### Merge and validation

Merged, and the post-merge validation passed.

<sub>atomic-foundry-status · foundry-run: 2026-07-30T15-13-42-256Z</sub>

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,116 @@
...
+        with:
+          fetch-depth: 0
+      - name: Install tools
+        uses: infrablocks/github-actions/asdf_install@v1
+      - name: Install secrets tools
+        run: ./scripts/ci/common/install-git-crypt.sh
+      - name: Unlock git-crypt
+        run: ./go git_crypt:unlock_with_encrypted_gpg_key
+        env:
+          ENCRYPTION_PASSPHRASE: ${{ secrets.ENCRYPTION_PASSPHRASE }}
```
phelma, line 74: 🟡 **Security** — [plan concern]

The PR `prerelease` job decrypts the git-crypt tree with `secrets.ENCRYPTION_PASSPHRASE` — exposing the Docker Hub push credentials and the CI GPG private key — while executing workflow steps and Rakefile code checked out from the *PR head branch*. Under `pull_request`, those are the PR's own versions, so anyone with write access to this repo (or an attacker who compromises such an account) can alter the Rakefile to exfiltrate the passphrase and the unlocked credentials. The old CircleCI pipeline ran secret-bearing jobs only on `main` and `dependabot/*`, so this broadens the blast radius to "anyone with write access holds publish rights over a widely consumed base image".

**Plan status**: this is D9, explicitly designed and documented — forks and dependabot are excluded by the `if`, the job is `contents: read`, it pushes only a PR-namespaced tag and never `latest`. Not a defect of this diff.

**For the plan, not this PR**: consider binding the secret to a protected `environment:`, or scoping a push-only PR-namespaced Docker Hub token, so "write access ⇒ publish credentials" becomes a stated decision rather than a side effect.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Unlock git-crypt
+        run: ./go git_crypt:unlock_with_encrypted_gpg_key
+        env:
+          ENCRYPTION_PASSPHRASE: ${{ secrets.ENCRYPTION_PASSPHRASE }}
+      - name: Set CI git author
+        run: ./go repository:set_ci_author
+      - name: Pull latest main
+        # Publishes main as of approval time, not the SHA that triggered the
+        # run; --ff-only so a non-fast-forward fails loudly
+        run: git pull --ff-only
```
phelma, line 133: 🟡 **Safety** — [plan concern]

The `release` job checks out `ref: main` and then `git pull --ff-only`, so it publishes whatever is on `main` at approval time — not the SHA `check`/`test` actually ran against. Approvals can sit for up to 30 days; anything merged in the interim is released untested by this run, and the approver's mental model ("I approved run X for SHA Y") does not match what ships.

**Plan status**: deliberate parity with the old `release.sh`, and the step carries a comment saying exactly this. No change required at cutover.

**For the plan, not this PR**: this is the highest-blast-radius inherited hazard in the diff — either drop the pull (release the tested SHA) or re-run `test` after the pull, as a fleet-wide follow-up.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,143 @@
...
+    # the approval window. queue: max keeps every queued run; the default
+    # cancels all but the newest.
+    concurrency:
+      group: main-tags
+      cancel-in-progress: false
+      queue: max
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
```
phelma, line 62: 🟡 **Correctness** — [plan concern]

With `queue: max` retaining every queued run, two close-together pushes to `main` can produce two approved `release` jobs. The first `version:release` turns the latest tag into a plain release; the second then evaluates `latest_tag.release!` against a version with no `pre` segment and `lib/version.rb` raises "no pre segment, this version is not a pre-release version" — after the job has already unlocked git-crypt and pulled main.

**Plan status**: the concurrency block is the plan's prescribed shape and `version`/`lib/version.rb` are untouched by design (D5), so this is not a defect of the diff.

**For the plan, not this PR**: making `version:release` a no-op (rather than a raise) when `latest_tag` is not a prerelease would turn a redundant approval into a benign outcome.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,143 @@
...
+    # No concurrency group here — a pending approval must not hold the publish slot
+    needs: [prerelease]
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    environment: release
+    steps:
+      - name: Assert release environment is protected
+        run: ./scripts/ci/common/assert-release-environment-protected.sh
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
phelma, line 101: 🔵 **Correctness / Safety** — [plan concern]

`release-gate` inherits the workflow-level `permissions: contents: read` and passes that `GITHUB_TOKEN` to `gh api repos/.../environments/release`. If that call ever fails on scope or reachability, `set -euo pipefail` fails the gate — the safe direction — but the operator sees the hand-written "no required-reviewers rule — run ./go pipeline:prepare" message rather than the real cause, and every release blocks until someone reads the raw `gh` output.

The plan itself records this caveat verbatim, and this YAML is prescribed, so it is not a defect of the diff.

**For the plan, not this PR**: capture `gh api`'s exit status separately and emit a distinct "cannot read environment (token scope?)" message, and assert the `reviewers` array is non-empty rather than only that the rule type exists.

Rakefile
```diff
@@ -248,3 +285,51 @@ namespace :version do
...
+             ".#{args.run_number}.#{args.run_attempt}"
+
+    Rake::Task['image:build'].invoke
+
+    credentials =
+      unless File.binread('config/secrets/dockerhub/credentials.yaml')
+                 .start_with?("\x00GITCRYPT")
+        YAML.load_file('config/secrets/dockerhub/credentials.yaml')
+      end
+    Docker.authenticate!(credentials) if credentials
```
phelma, line 308: 🔵 **Code Quality / Correctness** — [plan concern]

The git-crypt locked-tree detection exists in two independent copies with different shapes: the `dynamic do ... end` block in the `image` namespace (line ~238, which returns `nil` explicitly) and this inline `unless` form (which relies on `unless` returning `nil`). The credentials path and the `"\x00GITCRYPT"` sentinel are each duplicated. The two can drift, and this copy converts "tree is locked" into an opaque Docker Hub `denied` at push time rather than failing at the guard.

**Plan status**: both copies are the plan's literal template text (§4.4 steps 2a and 7), so this is not a defect of the diff.

**For the plan, not this PR**: a single `dockerhub_credentials` helper (with a `File.exist?` guard) used by both call sites would collapse the duplication and let `prerelease:publish` fail loudly on a locked tree.

Rakefile
```diff
@@ -248,3 +285,51 @@ namespace :version do
...
+  desc 'Build and publish a PR-namespaced pre-release image (PR CI only)'
+  task :publish, %i[pr_number run_number run_attempt] do |_, args|
+    # PR image tag: the version main's prerelease would publish next
+    # (latest_tag bumped by the family's prerelease type — normally rc),
+    # plus a collision-free per-run suffix in the family's tag grammar.
+    # Docker tags forbid '+', so we reuse the semver '-'/'.'-separated
+    # form and no build metadata. No git tag is created, pushed, or
+    # committed — unlike version:bump this only ever tags the built image.
+    pr_tag = "#{latest_tag.rc!}.pr#{args.pr_number}" \
+             ".#{args.run_number}.#{args.run_attempt}"
```
phelma, line 299: 🔵 **Correctness** — [plan concern]

`pr_tag` interpolates `args.pr_number`, `args.run_number` and `args.run_attempt` with no presence check. Rake task arguments are `nil` when omitted, so a bare `./go prerelease:publish` (or a workflow that loses one `env:` value) produces a tag like `0.5.0-rc.1.pr..` — still valid under Docker's tag grammar, so it is pushed successfully and each such run overwrites the previous one.

**Plan status**: this is the plan's prescribed task body verbatim, so it is not a defect of the diff.

**For the plan, not this PR**: a one-line argument presence check before building `pr_tag` would turn a silent malformed publish into an immediate failure.

Rakefile
```diff
@@ -142,51 +134,91 @@ namespace :library do
...
+  end
+end
+
+namespace :repository do
+  desc 'Set the git author for CI'
+  task :set_ci_author do
+    sh 'git config --global user.name "InfraBlocks CI"'
+    sh 'git config --global user.email "ci@infrablocks.io"'
+  end
 end
```
phelma, line 215: 🔵 **Safety** — [plan concern]

`repository:set_ci_author` runs `git config --global`, which is right on an ephemeral runner but destructive if a developer runs `./go repository:set_ci_author` locally — it permanently rewrites their machine-wide `user.name`/`user.email` to `InfraBlocks CI`, with no confirmation and no record of what it replaced. Subsequent commits in *every* repo on that machine are then attributed to CI.

**Plan status**: prescribed exactly as written (§4.4 step 5), and shared with the gem family, so it is not a defect of this diff.

**For the plan, not this PR**: guarding on `ENV['CI']`, or dropping `--global` in favour of repo-local config (sufficient for the runner's single checkout), would remove the local footgun fleet-wide.

Gemfile
```diff
@@ -7,17 +7,18 @@ gem 'git'
...
 gem 'rake_docker'
-gem 'rake_git'
-gem 'rake_git_crypt'
+gem 'rake_factory', '~> 0.33.0'
+gem 'rake_git', '~> 0.2.0'
+gem 'rake_git_crypt', '~> 0.4.0'
 gem 'rake_github'
 gem 'rake_gpg'
-gem 'rake_ssh'
+gem 'rake_slack'
```
phelma, line 16: 🔵 **Standards**

`rake_slack` is added without a version constraint, while every other gem added in this change (`rake_factory`, `rake_git`, `rake_git_crypt`, `ruby_git_crypt`) carries an explicit three-segment `~>` pin. The Rakefile depends on a specific `RakeSlack.define_notification_tasks` API shape (`bot_token` plus `routing_rules` with `when`/`channel`/`format` keys), which the lockfile currently resolves to 0.3.0, and the family plan itself describes the gem as `rake_slack (~> 0.3)`.

**Impact**: a future `bundle update` could pull a 0.4.x with a different routing-rules shape and break every Slack notify step, with nothing in the Gemfile signalling the expected line.

**Suggestion**: `gem 'rake_slack', '~> 0.3'`. This is the one finding in this review that is not a plan concern — it is a small, plan-neutral hardening of the diff.

phelma (commented):
## Code Review: #237 - Migrate pipeline from CircleCI to GitHub Actions

**Verdict:** COMMENT

Five lenses (correctness, security, safety, standards, code quality) reviewed this cutover. Measured against the Variant B family plan it is **fully conformant**: both workflow files reproduce the plan's prescribed YAML verbatim (job graph, one-command-per-step, `timeout-minutes`, `fetch-depth: 0` where tags matter, the job-level `main-tags`/`queue: max` concurrency with the slot-free `release-gate` split), the Rakefile changes land exactly as §4.4 specifies, the decommission is complete (`git grep -i circleci` matches nothing outside `Gemfile.lock`; `scripts/ci` retains only the three sanctioned scripts, all executable), the GPG key move is byte-identical (100% rename), `.tool-versions` is bumped to 3.3.11 with a settled lockfile (no prereleases, `ruby`/`x86_64-linux`/`x86_64-linux-gnu` platforms, no `rake_circle_ci`/`rake_ssh`), and the README key-management section is retitled and repointed. **No plan violations were found in either direction** — nothing required is missing, and nothing beyond the plan was changed.

Almost every finding below challenges something the plan or the PR description documents as deliberate. Those are marked **[plan concern]**: they are recorded so a human can revisit the *plan*, not this diff, and they do not block merge. Only one genuinely in-scope, plan-neutral point was found (the unconstrained `rake_slack` gem line).

### Cross-Cutting Themes

- **Secret exposure surface of the PR `prerelease` job** (flagged by: security, safety) — PR CI now unlocks git-crypt and pushes to the production Docker Hub repository while executing Rakefile/workflow code from the PR head branch. This is D9, explicitly designed and documented (fork/dependabot excluded, `contents: read`, PR-namespaced tag only, never `latest`), and the tag accumulation is an accepted cost. Worth an explicit fleet-wide trust statement post-migration. **[plan concern]**
- **Fail-quietly paths in the release chain** (flagged by: correctness, safety, code quality) — the locked-tree credentials guard, the missing-`SLACK_BOT_TOKEN` case, and the `assert-release-environment-protected.sh` error-message conflation all degrade silently or report the wrong cause. All three are the plan's prescribed shapes (and the plan itself calls out the last one). **[plan concern]**
- **Duplicated Rakefile knowledge** (flagged by: standards, code quality) — the git-crypt header check and the `alpine-aws` / `infrablocks/alpine-aws` identity strings each appear in two places. Both copies are the plan's literal template text. **[plan concern]**

### Tradeoff Analysis

- **Parity vs hardening**: security and safety would prefer scoped registry tokens, `persist-credentials: false`, update-type filters on dependabot auto-merge, and releasing the tested SHA rather than main-at-approval-time. The plan's §1 parity principle deliberately defers all of these to post-migration work. Recommendation: keep the diff as-is; carry these to the fleet-wide hardening backlog.
- **DRY vs flat CI**: code quality noted the seven-fold repetition of checkout/install/notify scaffolding and the ~45-line `prerelease:publish` body. The plan mandates the flat shape (logic lives in `./go`, CI stays lean) and prescribes the task body literally. Recommendation: no change; revisit via a composite action in `infrablocks/github-actions` only if the scaffolding grows.

### Strengths

- ✅ The `Docker::Image.get('alpine-aws:latest')` lookup was verified correct against the installed gem: `rake_docker` 2.21.0 builds with `options = { t: t.repository_name }`, so the bare-name/`:latest` assumption holds.
- ✅ The explicit `repo_tag:` on `image.push` avoids a real bug — docker-api would otherwise resolve the push repository from `RepoTags.first` and target `docker.io/library/alpine-aws`.
- ✅ `assert-release-environment-protected.sh` converts GitHub's fail-open auto-created environment into a fail-closed gate, and counts `required_reviewers` specifically rather than any protection rule. A genuine control that did not exist on CircleCI.
- ✅ No `pull_request_target`; attacker-controllable values (PR title, URL, number) pass through `env:` and are referenced as shell variables, closing the classic GHA script-injection vector.
- ✅ Least-privilege permissions throughout: top-level `contents: read`, `contents: write` only on main's prerelease/release, `contents: read` on the PR prerelease that must not push.
- ✅ `--match-head-commit "$HEAD_SHA"` closes the TOCTOU window between checks passing and the dependabot merge landing.
- ✅ Provisioning refuses to upload git-crypt ciphertext as `ENCRYPTION_PASSPHRASE`, preventing a corrupt secret that would only surface as an opaque GPG failure during a release.
- ✅ Every job carries an explicit `timeout-minutes`, bounding a hung compose-backed test that would otherwise hold a serialised slot for GitHub's 360-minute default.
- ✅ The git-crypt ciphertext detection is encoding-safe against `File.binread`'s ASCII-8BIT result and matches the 9-byte magic prefix correctly.
- ✅ Requires and gems stay strictly alphabetical; the long `RakeGithub` block uses the repo's existing `rubocop:disable Metrics/BlockLength` idiom rather than editing `.rubocop.yml`.

### False Positive (disproven, not counted)

- 🔵 **Security** flagged the pinned `docker/compose` `v5.3.1` release as implausible and its checksum as unverifiable. The plan records this pin as verified on 2026-07-20 against the release's own `checksums.txt`, with the `command -v` guard confirmed false on `ubuntu-latest` (24.04). No change warranted.

### General Findings

- 🔵 **Standards** (in-scope): `rake_slack` is added with no version constraint while the other gems added in this change carry three-segment `~>` pins, and the Rakefile is written against a specific `routing_rules` shape (resolved to 0.3.0). Consider `gem 'rake_slack', '~> 0.3'`. This is the only finding in the review that is not a plan concern.
- 🔵 **Correctness** [plan concern]: `latest_tag` returns `nil` on a tagless/shallow checkout, so `t.tags = [latest_tag.to_s, 'latest']` becomes `["", "latest"]` rather than erroring. Harmless today (only `check` is shallow, and it runs rubocop only), but the failure mode is silent. `lib/version.rb` and the `version`/`image` namespaces are untouched by design (D5).
- 🔵 **Correctness** [plan concern]: the Slack routing table is order-dependent — the `on_hold` rule must stay first and the `{ when: {} }` catch-all is terminal (anything below it is unreachable). Prescribed verbatim by the plan; a marker comment would make the ordering constraint self-documenting.
- 🔵 **Correctness** [plan concern]: `git pull --ff-only` in the release job relies on `actions/checkout` configuring branch tracking; `git pull --ff-only origin main` would be independent of that implementation detail.
- 🔵 **Safety** [plan concern]: no recovery guidance exists for the documented tag-then-publish hazard — when `image:publish` fails, an operator must know to delete the pushed tag before re-running. A README note would put that where the on-call reader finds it.
- 🔵 **Security** [plan concern]: the moved `.github/gpg.private.enc` still uses `openssl aes-256-cbc -md sha1 -k`, i.e. single-iteration `EVP_BytesToKey` with SHA-1, on a blob committed to a public repo. Inherited verbatim; a follow-up candidate (`-pbkdf2 -iter 600000 -md sha256`, or an Actions secret).
- 🔵 **Security** [plan concern]: deleting `config/secrets/{circle_ci,github}/config.yaml` does not revoke those credentials — explicitly a host-side sweep step, and correctly stated in the PR description. Ensure the sweep item has an owner and a deadline.
- 🔵 **Code Quality** [plan concern]: `prerelease:publish` mixes tag computation, build invocation, credential resolution, image lookup, re-tag and push in one ~45-line body; the `RakeGithub` block mixes imperative pre-flight validation into a configuration block, forcing the rubocop suppression. Both are the plan's literal prescribed text.
- 🔵 **Code Quality** [plan concern]: the prerelease bump type is hard-coded as `latest_tag.rc!` in the Rakefile while main's equivalent is parameterised as `version:bump[rc]` in the workflow — two places to change if the family's prerelease type ever moves.
- 🔵 **Code Quality** [plan concern]: neither Docker Hub credential read guards on `File.exist?`, unlike the passphrase read in the same file, so a clone without populated secrets fails with a bare `Errno::ENOENT` instead of an actionable message.

---

.github/workflows/pr.yaml
phelma: Accepted as designed — this doesn't broaden the blast radius.

`main` here has branch protection with only `allow_force_pushes: false` / `allow_deletions: false` — no required reviews, no required status checks, no rulesets. Anyone with write access can push straight to `main` and run the same job with the same secrets. The old pipeline also ran secret-bearing jobs on `dependabot/*` branches, which any writer can create. The security boundary is, and always was, **write access to the repo**; `pull_request` triggering doesn't move it.

What this does commit us to, stated once here rather than re-derived per repo: **write access implies git-crypt plaintext**, including for a collaborator never enrolled as a git-crypt GPG recipient. Enrolment is a convenience for local work, not an access control.

.github/workflows/main.yaml
phelma: Inherited from CircleCI — out of scope for the cutover.

At base `f687011`, `scripts/ci/steps/release.sh` runs a bare `git pull` before `version:release`, and `prerelease.sh` doesn't — the identical hazard *and* the identical prerelease/release asymmetry. The GHA form is marginally safer: `--ff-only` fails loudly where a bare `git pull` could merge and release a tree that never existed on `main`.

It is a genuine hardening candidate though, and the right fix is `ref: ${{ github.sha }}` on the release job rather than dropping the pull — `ref: main` is what opens the divergence window. Logged for the post-migration hardening sweep rather than a migration PR.

.github/workflows/main.yaml
phelma: Inherited — out of scope. `version:release` and `lib/version.rb` are untouched by this diff. CircleCI had *no* concurrency control at all, so two pushes raced simultaneously; the serialised queue is strictly better than what it replaces.

Gemfile
phelma: Yes, fixed in this repo separately, and added to the plan to make sure it doesn't happen again.

.github/workflows/main.yaml
phelma: This can't happen — the misleading message is unreachable on a `gh api` failure.

In `assert-release-environment-protected.sh`, `count=$(gh api …)` is a plain assignment, so its exit status is the command substitution's, `set -e` is active, and nothing masks it — no `local`, no `if` wrapper, no `&&` chain. A failing `gh api` exits the script *at the assignment*; the `if` is never reached and the hand-written "no required-reviewers rule" message is never printed. The operator sees `gh`'s own stderr, which is exactly what's being asked for here. The degenerate empty-`count` case fails the same way via `[ "" -eq 0 ]` raising "integer expression expected".

The second half — assert `reviewers` is non-empty rather than that the rule exists — is unreachable too: GitHub won't construct a `required_reviewers` rule with an empty reviewer list. Verified against the live environment, which carries the `maintainers` team.

Worth noting structurally: **the script is not the gate.** `release-gate` carries `environment: release`, so GitHub pauses for approval before any step runs. The script is a detector for "we got here without an approval" — i.e. the auto-created unprotected environment — and it catches that case correctly.

Rakefile
phelma: Real duplication, but only reachable on a local invocation.

The workflow unlocks git-crypt before `prerelease:publish`, and a failed unlock kills the job at that step — so the opaque Docker Hub `denied` at push time needs `./go prerelease:publish` run by hand on a locked local clone. Noisy rather than silent, and not something anyone does in anger. Not worth the change plus a family-wide re-run.

Rakefile
phelma: Same disposition as the git-crypt duplication above — local-invocation only.

`pr_number`/`run_number`/`run_attempt` are always populated in Actions, and the job is gated to same-repo non-dependabot PRs, so the pushable `…pr..` tag needs a bare local invocation to produce. Dismissed.

Rakefile
phelma: Inherited behaviour. `scripts/ci/common/configure-git.sh` already did `git config --global user.name` / `user.email` at base. Only the surface is new — it's now a rake task a developer could invoke locally. Dismissed.

phelma:
Walked the review-body findings as well (the ones without inline threads). Recording dispositions here so the same points don't get re-litigated across the remaining ~100 repos.

**Inherited from CircleCI — out of scope.** The cutover reproduces the old pipeline warts included; fixing a pre-existing hazard during migration is scope creep.

- **`latest_tag` nil → `["", "latest"]`** — `t.tags = [latest_tag.to_s, 'latest']` is byte-identical at base line 213.
- **No README tag-then-publish recovery note** — the tag-then-publish ordering predates the diff.
- **`gpg.private.enc` uses `-md sha1 -k`** — the file moved as a 100% rename; the `openssl aes-256-cbc -md sha1 -k` KDF is untouched.

**Docker Hub credentials lack a `File.exist?` guard** — mis-attributed. This was presented as an inconsistency introduced by the diff ("unlike the passphrase read in the same file"), but the unguarded `YAML.load_file` on `config/secrets/dockerhub/credentials.yaml` is pre-existing at base Rakefile line 207. Inherited, so out of scope.

**Deleted credentials are not revoked** — deferred deliberately. Removing `config/secrets/{circle_ci,github}/` takes the CircleCI API token and stored PAT out of the tree but leaves them live at their issuers and recoverable from history. CircleCI gets torn down wholesale once everything is migrated off it; revoking per-repo before then buys nothing and risks breaking a not-yet-migrated repo that shares credentials. It's on the end-of-migration teardown checklist.

**`git pull --ff-only` relies on branch tracking** — moot given the disposition on the `ref: main` thread above.

**Slack routing table order-dependence, `prerelease:publish` body mixing concerns, hard-coded `rc!` bump type** — dropped as style. Each costs a change plus a family-wide re-run and none can produce a wrong outcome.

One thing not in the review, surfaced while checking the `release` environment: it's provisioned with a `required_reviewers` rule naming the `maintainers` team, but `prevent_self_review` is `false` — so whoever merges to `main` can approve their own release, and combined with `main` having no required reviews, one person can take a change from commit to published `latest` with no second party. That's accepted: the gate exists to stop an *unattended* release, not a *unilateral* one, and requiring a second approver across ~100 repos would stall the fleet.

